PR #44, opened 2021-07-29
Correct `KEY` usage in `ATTACH`

m4heshd:
This PR fixes the minor incorrect documentation of `KEY` keyword usage in `ATTACH` SQL command. Currently documented usage will throw a syntax error (using `=`).


utelle:
Thanks for detecting and fixing this documentation glitch.

m4heshd:
> Thanks for detecting and fixing this documentation glitch.

My pleasure. 😊

PR #88, opened 2022-09-04
Wasm target

nullrocket:
Added conditional __WASM__ so that the entropy can be properly defined for wasm targets.


utelle (changes requested):
In principle, I'm not against adding additional code to support WebAssembly. However, I find your implementation a bit confusing.

1. Instead of using an `#elif` preprocessor instruction you are nesting `#if ... #endif` constructs. IMHO that makes the code less readable
2. There are quite a few unnecessary empty lines
3. Your code uses unconditionally console out (`printf`). Such output may be useful while debugging, but should not be in production code
4. Your implementation has actually unreachable code:
```c
#else
  #if defined(__WASM__)
    /* Code that will be executed if symbol __WASM__ is defined */
  #else
    #if defined(__WASM__)
      /* Code unreachable !!! */
      /* Symbol __WASM_ can't be defined here, because we are in*/
      /* the #else branch of the first check for symbol __WASM__ */
    #else
      # error "Secure pseudorandom number generator not implemented for this OS"
    #endif
#endif
```

I would prefer something like this (if that's what you want/need):
```c
#elif defined(__WASM__)
static size_t entropy(void* buf, size_t n)
{
  if (getentropy(buf, n) == 0)
    return n;
}
#else
# error "Secure pseudorandom number generator not implemented for this OS"
#endif
```

nullrocket:
Sorry about all the noise!  Something was wrong with my local git repo,  every time i commited changes, it would be a strange merge of an old commit with the new.  Fixed now.  I thought I was losing my mind!   

I do remember why I had the full braces and return here.
https://github.com/utelle/SQLite3MultipleCiphers/blob/cc7033eb6f88c68b5322e6fac3c1369aec554974/src/chacha20poly1305.c#L393

I was seeing this warning and thought I would clear it up. 

`warning: non-void function does not return a value in all control paths `

utelle (changes requested):
Thanks for cleaning up the code.

Unfortunately, I missed one glitch in the code; function `entropy` **must** return a value. That is, this code
```c
if (getentropy(buf, n) == 0)
    return n
```
has to be changed for example to
```c
  return (getentropy(buf, n) == 0) ? n : 0;
```

utelle:
> I do remember why I had the full braces and return here.
> 
> https://github.com/utelle/SQLite3MultipleCiphers/blob/cc7033eb6f88c68b5322e6fac3c1369aec554974/src/chacha20poly1305.c#L393
> 
> I was seeing this warning and thought I would clear it up.
> 
> `warning: non-void function does not return a value in all control paths `

I already commented on the latter, but additionally the keyword `static` should be dropped. That is, the final code for function `entropy` should look like this:

```c
extern size_t entropy(void* buf, size_t n);
size_t entropy(void* buf, size_t n)
{
  return (getentropy(buf, n) == 0) ? n : 0;
}
```

nullrocket:
Ok, I removed the static keyword,  just curious why?  There is another definition further up where I believe I got the signature from when I was making the change that is static.   

Also just to be sure, your code above is not what I have. 

> ```c
> extern size_t entropy(void* buf, size_t n);
> size_t entropy(void* buf, size_t n)
> {
>   return (getentropy(buf, n) == 0) ? n : 0;
> }
> ```

This is what I have. the extern decl is for getentropy.
```c
extern size_t getentropy(void* buf, size_t n);
size_t entropy(void* buf, size_t n)
{
  return (getentropy(buf, n) == 0) ? n : 0;
}
```

utelle:
> Ok, I removed the static keyword, just curious why? There is another definition further up where I believe I got the signature from when I was making the change that is static.

Ouch, please revert this change. That is, the `static` keyword should actually be used. Otherwise the symbol will be exported. Sorry for the confusion (it's late in the evening here in Germany 😉 ).

> Also just to be sure, your code above is not what I have.
> [...]
> This is what I have. the extern decl is for getentropy.
> 
> ```c
> extern size_t getentropy(void* buf, size_t n);
> size_t entropy(void* buf, size_t n)
> {
>   return (getentropy(buf, n) == 0) ? n : 0;
> }
> ```

Yes, your version is correct. Sorry for my lack of concentration.

utelle (approved):
I think the PR is now finally correct.

Issue #123, opened 2023-11-14
CMake: APIs not exported when building shared library


Myroendan:
Hi!
When building with CMake, the shared library is missing some important API-s (sqlite3_key for example). I have checked the CMakeLists.txt, there is a variable which is defined, but never used:
set(SQLITE3MC_DLLRES_SRCS
  src/sqlite3mc.def
  src/sqlite3mc.rc
)
The .def is needed to export the apis when building as a shared lib.
I added this to the sources, but then I got 54 undefined reference link errors. 
These 54 functions are not present in the static library, I'm not sure why they are listed. If I remove them from the def file, the shared lib builds fine and contains the necessary apis.

utelle:
> When building with CMake,

Unfortunately, I myself have only very limited experience with CMake. CMake support was contributed by another user.

> the shared library is missing some important API-s (sqlite3_key for example). I have checked the CMakeLists.txt, there is a variable which is defined, but never used: set(SQLITE3MC_DLLRES_SRCS src/sqlite3mc.def src/sqlite3mc.rc )

In case of a shared library build these files should be added, but obviously the CMake instruction to do that is missing.

> The .def is needed to export the apis when building as a shared lib. I added this to the sources, but then I got 54 undefined reference link errors.

As you observed the .def file is not really suited to solve the problem. The reason is that the .def file reflects the extension selection used by the pre-generated build files coming with SQLite3 Multiple Ciphers. Depending on the actual option selection in CMake the .def file will contain entries for functions that are not enabled.

> These 54 functions are not present in the static library, I'm not sure why they are listed.

They are listed, because they are compiled in with the pre-generated build files.

> If I remove them from the def file, the shared lib builds fine and contains the necessary apis.

A better option to enable the API actually compiled in will be to define the preprocessor symbol `SQLITE_API=__declspec(dllexport)` if a shared library/DLL is built.

I will try to adjust the CMake file, but this may take a while.

Myroendan:
Defining the preprocessor symbol worked, the .def file is not needed with this method. The .rc file should still be added to the sources if I'm correct. I will create a pull request with the changes.

utelle:
> Defining the preprocessor symbol worked, the .def file is not needed with this method.

Correct.

> The .rc file should still be added to the sources if I'm correct.

Yes, this will add version information to the resulting DLL.

> I will create a pull request with the changes.

TIA.

utelle:
[...]

PR #125, opened 2023-11-15
Add API exports when building shared lib with CMake

Myroendan:
- Fixed issue #123 - CMake: APIs not exported when building shared library

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -226,7 +226,6 @@ set(SQLITE3MC_BASE_SRCS
   src/test_windirent.h
 )
 set(SQLITE3MC_DLLRES_SRCS
-  src/sqlite3mc.def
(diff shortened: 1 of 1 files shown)


utelle (changes requested):
In principle, I'm fine with most changes. However, IMHO we will have to apply 2 additional (small) modifications:

1. Add the **linker flag** `-framework Security` if the library is built for **macOS** (system name `Darwin`) - this is required due to commit fb0b979c8ddba93a474e8ae9dbe41ab7b1496b0b in October.
2. Remove `sqlite3mc.def` from the `SQLITE3MC_DLLRES_SRCS` file list. Now, that the export definition is defined, this file is no longer needed.

Could you please apply these changes? TIA.

utelle (approved):
Thanks for applying the changes.

Issue #124, opened 2023-11-14
Conan package


Myroendan:
It would be nice to have a conan package available for this project. I've created a conan2 recipe, let me know if you would like to review it.

utelle:
> It would be nice to have a conan package available for this project. I've created a conan2 recipe, let me know if you would like to review it.

Admittedly, I haven't used _Conan_ up to now. However, it is certainly a good idea to support package management systems to make the component available for more users/developers.

So, yes, I'd like to review your _Conan_ recipe.

Myroendan:
I will create a PR for Issue#109 regarding cmake install first. If the build system handles the copy of the files (headers, lib etc..), the conan recipe will be shorter and cleaner.
I have one question to this topic. This is the list of headers which are added to the cmake target:
  src/cipher_common.h
  src/cipher_config.h
  src/fastpbkdf2.h
  src/mystdint.h
  src/rijndael.h
  src/sha1.h
  src/sha2.h
  src/sqlite3.h
  src/sqlite3ext.h
  src/sqlite3mc.h
  src/sqlite3mc_vfs.h
  src/sqlite3userauth.h
  src/test_windirent.h

Which ones shouldn't be used by users? I would only move the public ones to the include folder.

utelle:
> I will create a PR for Issue#109 regarding cmake install first.

Great. TIA.

> If the build system handles the copy of the files (headers, lib etc..), the conan recipe will be shorter and cleaner.

The "problem" with SQLite is that there are so many configuration options.

> I have one question to this topic. This is the list of headers which are added to the cmake target:

- [ ] `src/cipher_common.h`
- [ ] `src/cipher_config.h`
- [ ] `src/fastpbkdf2.h`
- [ ] `src/mystdint.h`
- [ ] `src/rijndael.h`
- [ ] `src/sha1.h`
- [ ] `src/sha2.h`
- [x] `src/sqlite3.h` - the include file of the original SQLite library
- [x] `src/sqlite3ext.h` - optional, required only when creating own SQLite extensions
- [x] `src/sqlite3mc.h` - **main include file** (includes all necessary header files)
- [x] `src/sqlite3mc_vfs.h` - public API of the encryption VFS
- [x] `src/sqlite3userauth.h` - optional, only used if user authentication extension is enabled
- [ ] `src/test_windirent.h`

> Which ones shouldn't be used by users? I would only move the public ones to the include folder.

[...]

PR #128, opened 2023-11-17
Add conan recipe

Myroendan:
- Closes issue #124: Conan package

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -213,6 +213,7 @@ set(SQLITE3MC_PUBLIC_HEADERS
   src/sqlite3.h
   src/sqlite3ext.h
   src/sqlite3mc.h
+  src/sqlite3mc_version.h
   src/sqlite3mc_vfs.h
   src/sqlite3userauth.h
 )
--- a/conan/all/test_package/main.c
+++ b/conan/all/test_package/main.c
@@ -1,0 +1,72 @@
+#include <sqlite3mc/sqlite3mc.h>
+#include <stdio.h>
+
+int main(int argc, char *argv[])
+{
+    sqlite3 *db;
+    const char *key = "password";
+    const char *wrongKey = "wrongPassword";
+
+    // Create database
+    int rc = sqlite3_open("test.db", &db);
+    if (rc != SQLITE_OK)
+    {
+        fprintf(stderr, "Failed to open database: %s\n", sqlite3_errmsg(db));
+        sqlite3_close(db);
+        return 1;
+    }
+
+    // Encrypt
+    sqlite3_key(db, key, strlen(key));
+
+    // Fill db with some data and close it
+    rc = sqlite3_exec(db, "CREATE TABLE users (name TEXT NOT NULL, ID INTEGER PRIMARY KEY UNIQUE)", NULL, NULL, NULL);
+    if (rc != SQLITE_OK)
+    {
+        fprintf(stderr, "SQL error: %s\n", sqlite3_errmsg(db));
+        sqlite3_close(db);
+        return 1;
+    }
+    rc = sqlite3_exec(db, "INSERT INTO users (name, ID) VALUES ('testUser', '12345')", NULL, NULL, NULL);
+    if (rc != SQLITE_OK)
+    {
+        fprintf(stderr, "SQL error: %s\n", sqlite3_errmsg(db));
+        sqlite3_close(db);
+        return 1;
+    }
+    rc = sqlite3_close(db);
+    if (rc != SQLITE_OK)
+    {
+        fprintf(stderr, "Failed to close database: %s\n", sqlite3_errmsg(db));
+        return 1;
+    }
+
+    // Reopen and provide wrong key
+    rc = sqlite3_open("test.db", &db);
+    if (rc != SQLITE_OK)
+    {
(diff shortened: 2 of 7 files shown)


CMakeLists.txt
```diff
@@ -352,7 +352,7 @@ if(SQLITE3MC_STATIC_RUNTIME_LINK)
 endif()
 message("Will build ${SQLITE3MC_TARGET} as ${SQLITE3MC_LINK}")
 
-set_target_properties(${SQLITE3MC_TARGET} PROPERTIES PUBLIC_HEADER "${SQLITE3MC_PUBLIC_HEADERS}")
+set_target_properties(${SQLITE3MC_TARGET} PROPERTIES PUBLIC_HEADER "${SQLITE3MC_PUBLIC_HEADERS};src/sqlite3mc_version.h")
```
utelle, line 355: We will have to make `sqlite3mc_version.h` part of the list `SQLITE3MC_PUBLIC_HEADERS`, because it is included by `sqlite3mc.h`. Sorry, my fault - I simply missed this include file when I flagged the public header files in the list in issue #124.

conan/all/test_package/main.c
```diff
@@ -0,0 +1,45 @@
...
+
+    int rc = sqlite3_open("test.db", &db);
+    if (rc != SQLITE_OK)
+    {
+        fprintf(stderr, "Failed to open database: %s\n", sqlite3_errmsg(db));
+        sqlite3_close(db);
+        return 1;
+    }
+
+    rc = sqlite3_key(db, key, strlen(key));
```
utelle, line 17: `sqlite3_key` can only fail, if it is called for an in-memory or temporary database, or if the URI specifies an invalid cipher scheme. It does **not** fail if an invalid password was given. A wrong password will be detected only, after the database file was actually accessed, that is, after any SQL DML command was issued.

For a newly created database file it will almost never fail.

utelle (changes requested):
There are still 2 small issues:

1. `sqlite3mc_version.h` has to be added to the list of public header files in `CMakeList.txt`
2. For me it is unclear what exactly the purpose of the test is. Since the test creates a new database file (at least, when it is called for the first time), it will never fail.

CMakeLists.txt
utelle: > This sqlite3mc_version.h is currently not added to the SQLITE3MC_TARGET as a source file.
> 
> * I can add it to SQLITE3MC_PUBLIC_HEADERS, but then it will be added as a source.

What's the problem with this approach? `sqlite3mc_version.h` **is** referenced from `sqlite3mc.h` and is therefore a **public** header file.

> * Or I can  append it to SQLITE3MC_PUBLIC_HEADERS, but only after the sources were added to SQLITE3MC_TARGET.

Why?

>       Which one should be?

IMHO it should be **added** to `SQLITE3MC_PUBLIC_HEADERS`.

Currently you have defined `SQLITE3MC_PUBLIC_HEADERS` as
```
set(SQLITE3MC_PUBLIC_HEADERS
  src/sqlite3.h
  src/sqlite3ext.h
  src/sqlite3mc.h
  src/sqlite3mc_vfs.h
  src/sqlite3userauth.h
)
```
because these are the header files a user of the library will need to be able to compile and build an application. `sqlite3mc_version.h` is included by `sqlite3mc.h` - that is, if it is missing compiling an application will fail. Therefore I think the right place to add it is `SQLITE3MC_PUBLIC_HEADERS`.

Myroendan: Alright I'll add it to the public headers.
I was just wondering how did this compile so far. But I guess it's because the sqlite3mc_version.h was only used by sqlite3mc.h, and none of the other source files.

conan/all/test_package/main.c
Myroendan: I have tried the following:
1. Creating a db
2. Use sqlite3_key
3. Creating a users table and insert one user
4. Close the db
5. Open the db again
6. Use sqlite3_key with a wrong password

But it's returning with SQLITE_OK. Am I missing something?
If I try to access an already existing encrypted db with another exe, its working, I can't access it without the correct password. But I can't reproduce the same within one test program.

CMakeLists.txt
utelle: > Alright I'll add it to the public headers.

Thanks.

> I was just wondering how did this compile so far. But I guess it's because the sqlite3mc_version.h was only used by sqlite3mc.h, and none of the other source files.

Well, `sqlite3mc.c` includes `sqlite3mc.h` which in turn includes `sqlite3mc_version.h`, and for example the version string (`SQLITE3MC_VERSION_STRING`) is used in function `sqlite3mc_version()` of `sqlite3mc.c` to return the version information. It compiles, because the header file resides in the same directory as the other header files and the source files.

conan/all/test_package/main.c
Myroendan: > You mean sqlite3_key returns an error??? That can't be. An error is returned when you try to read or write data.

You're right, the query returned the error, not the sqlite3_key.

> Just issue a query after calling sqlite3_key. This should result in an error if the wrong password was provided.

Thanks, it's working now. I'll update the PR this evening.

utelle:
Thanks for addressing the open questions. I will review the changes tomorrow.

utelle (approved):
Most likely the test application could be further improved, but this can be done later on, if necessary. For now, I will approve the PR.

PR #134, opened 2023-11-24
Add SQLITE_API for user authentication functions

taozuhong:


utelle (approved):
Thanks for the adjustment of the function signatures for the user authentication extension.

PR #147, opened 2024-03-31
Fix Secure Memory

jammerxd:
Fixed a missing preprocessor definition and missing semi-colon.

SQLITE3MC_USE_RANDOM_FILL_MEMORY and SQLITE3MC_ENABLE_RANDOM_FILL_MEMORY are required

There's a missing semi-colon that makes the compiler error.


utelle:
> Fixed a missing preprocessor definition and missing semi-colon.
> 
> SQLITE3MC_USE_RANDOM_FILL_MEMORY and SQLITE3MC_ENABLE_RANDOM_FILL_MEMORY are required

Well, actually the symbol `SQLITE3MC_ENABLE_RANDOM_FILL_MEMORY` is used only once in the source. And I think it should be replaced by `SQLITE3MC_USE_RANDOM_FILL_MEMORY`. That's something I obviously forgot to do ... indicating that I'm not convinced myself whether to expose this feature at all.

IMHO it is enough to set unallocated memory to zero values. Using random data to fill unallocated memory just incurs additional overhead without increasing security. Therefore I tend to remove this option completely.

> There's a missing semi-colon that makes the compiler error.

Yes, correct. But it will no longer matter if the _random fill_ option is removed.

If you think the option to fill unallocated memory with random data should be kept, please let me know your reasoning.

jammerxd:
Gotcha - it's just that the problem right now is if you set SQLITE3MC_USE_RANDOM_FILL_MEMORY - you get 2 compile errors - one from the missing semi-colon, the other from the function that isn't exported (unresolved external symbol compile error). If you're planning on removing that option - that's okay, but we should still make it so it compiles in the meantime.

utelle:
> Gotcha - it's just that the problem right now is if you set SQLITE3MC_USE_RANDOM_FILL_MEMORY - you get 2 compile errors - one from the missing semi-colon, the other from the function that isn't exported (unresolved external symbol compile error).

Yes, that's correct. But only if you explicitly choose to use this option (which is off by default).

> If you're planning on removing that option - that's okay, but we should still make it so it compiles in the meantime.

Commit 80722b01faeb96328579c5cd594a722240040c7f removes the compile time option `SQLITE3MC_USE_RANDOM_FILL_MEMORY `. Therefore I will close this PR.

Nevertheless thank you very much for reporting the issue.

PR #181, opened 2024-11-17
Added compilation params for Android

FalsinSoft:
Without these params doesn't compiler for Android using NDK cause aes support missing


utelle:
The next version of _SQLite3 Multiple Ciphers_ hopefully doesn't need those extra compile flags any longer. However, I'm not absolutely sure whether setting the required compile options in the sources works on Android platforms. Therefore I accept the PR for now. Thanks.

PR #184, opened 2024-12-10
Fix undefined behavior related to function pointer cast

Koromix:
Casting a function pointer to a different function pointer type is undefined behavior in C, and triggers UBSan as shown below:

image

This PR fixes this by changing the declaration of `sqlite3mcFreeCodecParameterTable` to take a `void *` pointer and avoid the cast entirely.


utelle:
Thanks for reporting the issue and providing a fix.

PR #203, opened 2025-05-17
Make it available for find_package()

Pichas:
For now when you run `cmake --target install` it copies only headers and lib files, but fails if you try to use `find_package(sqlite3mc)` 

This PR adds cmake code to generate `sqlite3mcConfig.cmake`, `sqlite3mcConfigVersion.cmake`, `sqlite3mcTargets.cmake`. So now you can compile and install it into linux system and use `find_package()` to compile another projects with this library. All other libraries which uses this lib can be exported and installed without `target "sqlite3mc" that is not in any export set` error.

This PR was also tested as conan package and git submodule.

Example for static lib. Shared version works too.

```bash
# execute in the root folder to compile and install
cmake -D SQLITE3MC_STATIC=ON -Bbuild -S. --fresh
cmake --build ./build --target install
```

```cmake
# cmake file for tests
cmake_minimum_required(VERSION 3.20...4.0)

project(my_test)

add_executable(${PROJECT_NAME} main.cpp)

find_package(sqlite3mc_static REQUIRED)
target_link_libraries(${PROJECT_NAME} PRIVATE sqlite3mc_static)
```

```cpp
// main.cpp
#include <sqlite3mc/sqlite3mc.h>
#include <stdio.h>

int main() {
  printf("cipher_count: %d\n", sqlite3mc_cipher_count());
  return 0;
}
```


CMakeLists.txt
```diff
@@ -401,14 +432,13 @@ if(SQLITE3MC_BUILD_SHELL)
 
   if(SQLITE3MC_RUNTIME_LINK)
     message("Will build ${SQLITE3MC_SHELL_TARGET} with static runtime")
-    _Enable_MT(${SQLIT3MC_SHELL_TARGET})
+    _Enable_MT(${SQLITE3MC_SHELL_TARGET})
```
Pichas, line 435: fixed typo here

utelle:
> For now when you run `cmake --target install` it copies only headers and lib files, but fails if you try to use `find_package(sqlite3mc)`
> 
> This PR adds cmake code to generate `sqlite3mcConfig.cmake`, `sqlite3mcConfigVersion.cmake`, `sqlite3mcTargets.cmake`. So now you can compile and install it into linux system and use `find_package()` to compile another projects with this library. All other libraries which uses this lib can be exported and installed without `target "sqlite3mc" that is not in any export set` error.

Sounds great. Thank you for your efforts!

> This PR was also tested as conan package and git submodule.

Fantastic. This improvement of cmake support will certainly be very welcome by many users of SQLite3MC. I will merge your PR soon.

PR #210, opened 2025-07-18
Use a single name for cmake find_package

jcar87:
The PR that introduced the `find_package` capabilities does not follow usual CMake conventions, namely:
- the package has two different names, depending on whether shared or static was built
- the exported targets are not prefixed with a cmake namespace

this PR introduces this, such the config and targets work for the following:
* `find_package(sqlite3mc)` works
*  `target_link_libraries(xxx PRIVATE sqlite3mc::sqlite3mc_static)`
* OR ``target_link_libraries(xxx PRIVATE sqlite3mc::sqlite3mc)``

This is in line with current CMake convention (one may want to consider having the `sqlite3mc::sqlite3mc` be an alias for the static recipe if the shared one is not available). 

By all means, feel free to disregard this PR altogether - it's just something that I noticed while reviewing https://github.com/conan-io/conan-center-index/pull/27924. Lack of a namespace is relatively common (but not recommended), but the package having different names is rather unusual.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -360,7 +360,7 @@ message("Will build ${SQLITE3MC_TARGET} as ${SQLITE3MC_LINK}")
 
 include(GNUInstallDirs)
 INSTALL(TARGETS ${SQLITE3MC_TARGET}
-  EXPORT ${SQLITE3MC_TARGET}Targets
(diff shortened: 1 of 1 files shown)


utelle:
@jcar87 Thank you very much for providing these improvements to the CMake build.

PR #254, opened 2026-09-03
chacha20: Reorder MAC verification and decryption process

matbech:
## Description

Refactor decryption logic to verify MAC before decrypting data. In addition, it removes the unused allzero variable and chacha20_ismemset check.

## Problem Statement

- verify MAC before decrypting data
- remove unused allzero variable and check
 
## Type of Change

- [x] Refactor

## Checklist

### DIFF
--- a/src/cipher_chacha20.c
+++ b/src/cipher_chacha20.c
@@ -311,19 +311,16 @@ DecryptPageChaCha20Cipher(void* cipher, int page, unsigned char* data, int len, 
 
   if (nReserved > 0)
   {
-    int allzero = 0;
     /* Decrypt and verify MAC */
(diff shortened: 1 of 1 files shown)


utelle:
> Refactor decryption logic to verify MAC before decrypting data.

For normal operation the order doesn't matter, but in the event of an error it makes sense. Most likely the same logic should be applied to other cipher schemes with MACs.

> In addition, it removes the unused allzero variable and chacha20_ismemset check.

Most likely I was distracted somehow back then when I added this test. AFAICR a special case can occur where the page is allzero, and maybe that should be handled explicitly. I have to dig in my notes to be sure.

PR #255, opened 2026-09-03
Allow value-level encryption to be controlled by a compiler define

matbech:
## Description

Allow value-level encryption to be controlled by a compiler define. Preserve enabled default unless SQLITE3MC_ENABLE_VLE is supplied by the build.

## Type of Change

- [x] Refactor (no behavior change)

## Problem / Motivation

For Refactors:
Allow value-level encryption to be controlled by a compiler define.

## Checklist

### DIFF
--- a/src/sqlite3mc.c
+++ b/src/sqlite3mc.c
@@ -370,8 +370,10 @@ int sqlite3_zipfile_init(sqlite3 *db, char **pzErrMsg, const sqlite3_api_routine
 /*
 ** VLE - Value Level Encryption
 */
+#ifndef SQLITE3MC_ENABLE_VLE
 #define SQLITE3MC_ENABLE_VLE 1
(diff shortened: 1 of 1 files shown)


utelle:
> Allow value-level encryption to be controlled by a compiler define. Preserve enabled default unless SQLITE3MC_ENABLE_VLE is supplied by the build.

Actually, the unconditional `#define SQLITE3MC_ENABLE_VLE 1` remained there accidentally. I simply forgot to remove it after my first experiments with the new feature. The idea was to only enable the VLE feature if it is explicitly requested.

So, it boils down to the question whether the VLE feature should be enabled by default or not.

utelle:
I removed the unconditional define for _SQLITE3MC_ENABLE_VLE_. To have the feature disabled by default is the intended behaviour. And now the feature can be enabled by the user's descretion.

PR #258, opened 2026-09-03
chacha20poly1305: Implement SIMD optimizations for tag comparison

matbech:
## Description

Add SIMD support for poly1305 tag comparison.

## Type of Change

- [x] Feature / Enhancement (new functionality)

## Problem / Motivation
Improves performance for WASM SIMD and -msse4.1 or /arch:SSE4.2 with MSVC.

## Checklist

### DIFF
--- a/src/chacha20poly1305.c
+++ b/src/chacha20poly1305.c
@@ -10,6 +10,12 @@
 #include "mystdint.h"
 #include <string.h>
 
+#if defined(__wasm_simd128__)
+#include <wasm_simd128.h>
+#elif defined(__SSE4_1__)
+#include <immintrin.h>
+#endif
+
 #define ROL32(x, c) (((x) << (c)) | ((x) >> (32-(c))))
 #define ROR32(x, c) (((x) >> (c)) | ((x) << (32-(c))))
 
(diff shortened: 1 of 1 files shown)


utelle:
In its current form I can't accept this PR unfortunately. For _SQLite3MC_ I chose the approach to instruct the compiler by setting target attributes, which CPU features to use. But which code branch is actually used is decided at runtime by checking the available CPU features.

Your approach requires to specify the correct compile options, otherwise the fallback branch is used, even if the CPU is able to excute better variant.

Since the amount of code is small, I can handle the implementation in SQLite3MC's style.

matbech:
I could remove the implementations for the x86 and aarch64 architectures, which would only leave WASM SIMD. Would that be acceptable?

utelle:
> I could remove the implementations for the x86 and aarch64 architectures, which would only leave WASM SIMD.

That's not the point. Of course, it makes sense to make use of available CPU features and to accelerate the implementation for the x86 and aarch64 architecture, too.

I already started to code it in the way that the best available implementation is chosen at runtime.

Yes, I know it's different for WASM, because feature detection at runtime is not supported. So, for WASM using SIMD can only achieved by using the corresponding compile time option.

matbech:
I agree. Architecture-specific implementations should be separated into their own files. I'll close this PR for now, but feel free to open a new one with the restructured approach.

utelle:
> I agree. Architecture-specific implementations should be separated into their own files. I'll close this PR for now, but feel free to open a new one with the restructured approach.

I added the architecture-specific `poly1305_tagcmp` implementations in branch `hwaccel` to make sure I didn't introduce any bugs. The CI runs succeeded, so that the changes can be applied to the `main` branch.

PR #262, opened 2026-09-11
Use AES hardware support on Apple arm64 and with GCC on aarch64

mtrossbach:
Hi Ulrich,

first of all, thank you for SQLite3 Multiple Ciphers! It's a great piece of work, and the modular cipher design made it easy to find my way around the code.

While measuring AES performance on Apple Silicon, I noticed two gaps in `src/aes_hardware.c`:

- The existing ARMv8 AES code is compiled on Apple devices but never used.
- GCC builds on aarch64 don't include it at all.

This small PR closes both. The encryption output doesn't change in any way; the only difference is that the much faster hardware path is now picked wherever it's available.

### What changes (only `src/aes_hardware.c`, +23 lines)

- **Apple arm64 (macOS, iOS, iPadOS, tvOS, watchOS, visionOS):** `aesHardwareAvailableOnPlatform()` had no Apple branch and fell through to `return 0`. As a result, `aes128cbc`, `aes256cbc` and `sqlcipher` used the table-based software AES. All 64-bit ARM processors Apple uses (A7 and later, M1 and later) implement the ARMv8 cryptographic extensions, so the new branch simply returns 1. OpenSSL and Botan make the same assumption, and the bundled libaegis already relies on it (`src/aegis/common/cpu.c`).
- **GCC on aarch64:** the compiler detection only covered GCC on x86, so without `-march=...+crypto` no hardware path was compiled. For GCC ≥ 6, the NEON code path is now enabled with the function attribute `target("+crypto")`, just like the existing clang handling. The runtime check (`getauxval(AT_HWCAP) & HWCAP_AES`) was already in place.
- **Everything else is untouched:** x86, MSVC, clang, Windows on ARM, and the CBC functions themselves.

A nice side effect: the existing CI jobs on `macos-latest` and `ubuntu-24.04-arm` (gcc) will now really exercise the hardware code paths. At the moment they quietly test the software fallback.

### Performance (Apple M4 Max)

- **AES-256-CBC, 4 KiB blocks:** encrypt 314 → 2145 MB/s, decrypt 386 → 16392 MB/s.
- **`aes256cbc` database, 100,000 rows × 1,000 bytes:**
  - Write: 0.59 s → 0.30 s.
  - Read: 0.31 s → 0.06 s.
  - On Linux aarch64 with GCC 13, read time goes from 0.33 s → 0.08 s.

<details>
<summary>How I tested it</summary>

- **macOS (Apple M4 Max, Apple clang 21):**
[...]

PR #266, opened 2026-09-13
Add tests for the cryptographic primitives

mtrossbach:
Hi Ulrich,

as discussed in #262, this adds a test program for the cryptographic primitives and runs it in CI.

**Type:** Feature / Enhancement (tests)

**What it checks**

`test/cryptotest.c` includes the amalgamation, so it can call the internal functions directly:

- AES-CBC with the NIST SP 800-38A vectors: the software core, the Rijndael CBC functions used by the cipher schemes and, if available, the hardware path,
- ChaCha20 and Poly1305 with the RFC 8439 test vectors, for every implementation that is compiled for the target and supported by the CPU,
- that all ChaCha20 and Poly1305 implementations return the same results for pseudo-random input of many lengths,
- that all tag compare implementations detect equal tags and tags that differ in a single bit.

Without selectable implementations (currently `main`), only the default implementations are checked. The program returns 1 if a check fails.

Because the program calls internal functions such as `sse2_poly1305()` directly, it has to be adjusted when these functions are renamed or removed.

**Why**

The existing CI tests only exercise the implementations a runner selects by default. A hardware-specific implementation that computes wrong results, or results that differ from the other implementations, would not be noticed.

**Build and CI**

`Makefile.am` adds `cryptotest` as a `check_PROGRAMS` target with the flags of `sqlite3shell`, so a normal `make` doesn't build it. The `host_unix` job builds and runs it after the existing tests.

**Tested**

- `main` (a393e13): the CI recipe (`autoreconf`, `./configure`, `make`, then the new step) on Linux x86_64 and arm64 with GCC and Clang, plus direct builds on macOS arm64, the iOS simulator, Android (NDK r28) and WebAssembly. All checks pass.
- `hwaccel` (8794b10): all 6669 checks pass on ARM64, including NEON, and all 9123 checks pass on x86_64, including SSSE3, AVX2 and SSE2.

The test vectors were cross-checked with OpenSSL.

Best regards,
Markus


utelle:
Hi Markus,

thank very much for this great contribution! 🥇

Issue #148, opened 2024-04-19
Encrypting temporary files


ncruces:
Hi. I'm trying to develop my own encrypting VFS (n the context of developing a [Go driver](https://github.com/ncruces/go-sqlite3/)). I was looking into your code for inspiration, and have a few comments/concerns.

Have you tested this branch?
https://github.com/utelle/SQLite3MultipleCiphers/blob/80722b01faeb96328579c5cd594a722240040c7f/src/sqlite3mc_vfs.c#L435-L441

I… don't think you can do this: [`sqlite3_filename_database`](https://www.sqlite.org/c3ref/filename_database.html) is only safe to call with a *main* database, journal, or WAL name? None of the other kinds of files are safe to call this, and can corrupt memory…

ncruces:
That said, encrypting these files is important.

`SQLITE_OPEN_SUBJOURNAL` is used for [`SNAPSHOT`](https://www.sqlite.org/c3ref/snapshot.html)s, and [statement journals](https://www.sqlite.org/tempfiles.html#statement_journal_files).

`SQLITE_OPEN_TRANSIENT_DB` are (I think) used for [materialization of queries](https://www.sqlite.org/tempfiles.html#materializations_of_views_and_subqueries), [transient indices](https://www.sqlite.org/tempfiles.html#transient_indices) and [transient databases for VACUUM](https://www.sqlite.org/tempfiles.html#transient_database_used_by_vacuum_).

`SQLITE_OPEN_TEMP_JOURNAL` are used for journals of `TEMP` databases, but are _also_ used by the VBDE sorter (in a format called PMA that's nothing like a journal).

So all of these potentially store database data, though not necessarily in the form of pages (and with reserved bytes at the end).

utelle:
> Have you tested this branch?
> 
> https://github.com/utelle/SQLite3MultipleCiphers/blob/80722b01faeb96328579c5cd594a722240040c7f/src/sqlite3mc_vfs.c#L435-L441
> 
> I… don't think you can do this: [`sqlite3_filename_database`](https://www.sqlite.org/c3ref/filename_database.html) is only safe to call with a _main_ database, journal, or WAL name? None of the other kinds of files are safe to call this, and can corrupt memory…

Up to now I haven't experienced any memory corruption, and I haven't received any bug reports either. However, it can indeed happen that `zName` is a NULL pointer. I have to further investigate how to handle that situation.

> That said, encrypting these files is important.
> 
[...]

PR #271, opened 2026-09-15
Keep temporary data in memory unless SQLITE_TEMP_STORE is set

mtrossbach:
Hi Ulrich,

a small follow-up to #148.

**The problem**

Temporary files (temporary tables, sorter files, transient databases) are not encrypted, as discussed in #148. The documentation therefore strongly recommends `SQLITE_TEMP_STORE=2` or `3`.

The build files of SQLite3 Multiple Ciphers set `SQLITE_TEMP_STORE=2`, but the source itself doesn't define it. So the setting doesn't apply when the amalgamation (or `src/sqlite3mc.c`) is compiled with other build scripts, as many packages do. Without the flag, SQLite's default `1` applies, and temporary data is written to disk in plaintext. Of ten projects I looked at that compile SQLite3 Multiple Ciphers themselves, four don't set it; one of them sets `PRAGMA temp_store=MEMORY` at runtime instead.

**The change**

If `SQLITE_TEMP_STORE` is not defined, `sqlite3mc.c` now defines it as `2` before including SQLite. The amalgamation then behaves the same as the builds from the build files, whatever it is built with. An explicit value, including `1`, is kept, and `PRAGMA temp_store` works as before.

**What do you think?**

Would this be a good idea for you? SQLCipher goes further: since March 2025 its build stops unless `SQLITE_TEMP_STORE` is `2` or `3`. I chose the softer default so that builds that set `1` on purpose keep working. If you'd rather have an error like that, or prefer to leave this to the build files as before, that's fine with me too.

<details>

I built the shell from `src/sqlite3mc.c` without setting `SQLITE_TEMP_STORE`, filled a temporary table with a marker and searched the open temporary files of the process for it (Linux, `/proc/<pid>/fd`):

| Build | `TEMP_STORE` | Marker in plaintext in a temporary file |
|---|---|---|
| `main` (8392d70), no flag | 1 | yes |
| with this change, no flag | 2 | no |
| with this change, `-DSQLITE_TEMP_STORE=1` | 1 | yes (the explicit value is kept) |
| with this change, `-DSQLITE_TEMP_STORE=3` | 3 | no |
| amalgamation from `scripts/amalgamate.py` with this change, no flag | 2 | no |

The build files already set `SQLITE_TEMP_STORE=2`, so builds made with them don't change.

</details>

Best regards,
Markus


utelle:
Hi Markus,

> **The problem**
> 
> Temporary files (temporary tables, sorter files, transient databases) are not encrypted, as discussed in #148. The documentation therefore strongly recommends `SQLITE_TEMP_STORE=2` or `3`.

Correct. Unfortunately, I haven't found a (reasonable) way to force encryption for temporary SQLite files in case of an encrypted main database.

> The build files of SQLite3 Multiple Ciphers set `SQLITE_TEMP_STORE=2`, but the source itself doesn't define it. So the setting doesn't apply when the amalgamation (or `src/sqlite3mc.c`) is compiled with other build scripts, as many packages do. Without the flag, SQLite's default `1` applies, and temporary data is written to disk in plaintext. Of ten projects I looked at that compile SQLite3 Multiple Ciphers themselves, four don't set it; one of them sets `PRAGMA temp_store=MEMORY` at runtime instead.

I fully agree that this is anything but ideal. And unfortunately, few people read the documentation. So, it seems one has to force people to be happy sometimes.

> **The change**
> 
> If `SQLITE_TEMP_STORE` is not defined, `sqlite3mc.c` now defines it as `2` before including SQLite. The amalgamation then behaves the same as the builds from the build files, whatever it is built with. An explicit value, including `1`, is kept, and `PRAGMA temp_store` works as before.

Good idea. And it should have been applied much earlier. But better late than never.

> Would this be a good idea for you?

Yes, absolutely.

> SQLCipher goes further: since March 2025 its build stops unless `SQLITE_TEMP_STORE` is `2` or `3`. I chose the softer default so that builds that set `1` on purpose keep working.

If the option was set intentionally, it should be respected.

> If you'd rather have an error like that, or prefer to leave this to the build files as before, that's fine with me too.

No, I think it is a good thing to set `SQLITE_TEMP_STORE` to a reasonable default, if it wasn't set explicitly.

However, I would prefer to have these preprocessor statements further up in the source file `sqlite3mc.c`.  For example, directly at the beginning of the file (after line 9). There might be other options now or in the future that should deviate from the usual SQLite defaults. One should not have to search their settings in the depths of the source file.

mtrossbach:
Hi Ulrich,

thanks! I moved the block to the beginning of `sqlite3mc.c`, right after the file header, and added the meaning of the values to the comment. The tests from the description give the same results.

Best regards,
Markus

utelle:
Hi Markus,

> I moved the block to the beginning of `sqlite3mc.c`, right after the file header, and added the meaning of the values to the comment.

Perfect! Thanks!

